PR #17485, opened 2026-08-21
Template escaping is a no-op, so an apostrophe emits unparseable code

### LINKED ISSUE #17484: Template escaping is a no-op, so an apostrophe emits unparseable code
## Context

Operator surfaced the [code-scanning queue](https://github.com/neomjs/neo/security/code-scanning) for triage. Two of the open alerts — **41** and **42**, rule `js/identity-replacement` — are a genuine defect rather than a false positive, and this ticket carries only those two. The `src/Neo.mjs` prototype alerts (62/63) are deliberate prototype *enrichment* and want a documented dismissal, not a fix; alert 64 is a separate rule in a separate file. Neither is in scope here.

**This was verified before filing, not inferred from the alert.** CodeQL named the line; the reproduction below is mine.

Live latest-open sweep: checked latest 20 open issues at 2026-08-21T15:5x UTC; `gh search issues` for `identity-replacement OR templateBuildProcessor` returned zero; corpus grep over `resources/content/issues/**` for `templateBuildProcessor` returned zero; A2A claim sweep over the last 30 messages found no overlapping lane. No equivalent exists.

## The Problem

Both sites intend to escape single quotes before wrapping a static chunk in a single-quoted JS string literal. The comment at `:119` says so outright:

```js
// Escape single quotes for the string literal part of the chain.
return `'${part.replace(/'/g, "\'" )}'`;
```

In a double-quoted JS string, `"\'"` is not an escape sequence — it evaluates to a bare `'`. The call therefore replaces every apostrophe **with itself**. The intent and the behaviour have diverged silently since the line was written, and nothing downstream notices because the failure needs an apostrophe in static template text *adjacent to a dynamic part*.

**Proven, not asserted:**

```
input : "it's here"
output: "it's here"
IDENTITY: true
```

**Reproduced through the real exported function**, not a reconstruction of it — `processHtmlTemplateLiteral(['<div>it\'s ', ' here</div>'], ['count'])`, i.e. the build-time equivalent of ``html`<div>it's ${count} here</div>` ``:

```json
{"tag":"div","text":"##__NEO_EXPR__'it's ' + (count) + ' here'##__NEO_EXPR__##"}
```

The emitted expression is `'it's ' + (count) + ' here'`. **With a positive control:**

```
SYNTAX ERROR: Unexpected identifier 's'
CONTROL (escaped) PARSES OK
```

The control matters — it establishes that the correctly-escaped form parses, so the failure is attributable to the escaping and not to the surrounding expression shape.

**Realistic impact is build-time breakage, not a runtime vulnerability.** Template text is author-controlled source, so this is a correctness defect that happens to wear an injection-shaped rule name: any component author who writes a contraction in static markup next to an interpolation gets generated code that will not parse. The `security` label is here so the alert-triage path finds the ticket, not as a severity claim.

## The Architectural Reality

`buildScripts/util/templateBuildProcessor.mjs` converts tagged `html` template literals into Neo VDOM at build time. Static chunks and `##__NEO_EXPR__` placeholders are joined into a single `+` addition chain, which is later emitted as source. The two flawed sites are the only places that construct a quoted literal from a raw static chunk:

- `buildScripts/util/templateBuildProcessor.mjs:120` — the text-node chain
- `buildScripts/util/templateBuildProcessor.mjs:182` — the attribute-value chain

Both are reached from the sole export, `processHtmlTemplateLiteral` at `:312`. The two paths are independent: `:120` handles children, `:182` handles attributes, so a fix to one does not cover the other and each needs its own coverage.

## The Fix

Replace the identity replacement with a real escape at both sites, so the emitted literal carries a backslash:

```js
part.replace(/'/g, "\\'")
```

A backslash in the static chunk must also survive, which the current code does not handle either — escape backslashes before quotes, or the emitted literal mis-nests on input containing `\`. The fix should handle both characters in the correct order rather than only the character CodeQL named.

Prefer routing both sites through one shared helper rather than repeating the expression twice: the defect exists in duplicate precisely because the logic was inlined at two call sites, and a third chain would inherit it again.

## Acceptance Criteria

- [ ] `:120` and `:182` emit a correctly-escaped single-quoted literal; both go through one shared escaping helper rather than duplicated inline expressions
- [ ] Backslash in static text is escaped before the quote, so `\` and `'` in the same chunk both survive
- [ ] RED-first: a test asserts the emitted expression **parses** (not merely that it string-matches) and fails on current `dev` for the text-node path
- [ ] A second RED arm covers the attribute path at `:182` independently — a fix to one site must not green the other's arm
- [ ] Mutation control: reverting either site individually reddens its own arm and only its own arm
- [ ] CodeQL alerts 41 and 42 close on the merge commit

## Out of Scope

- Alerts **62/63** (`js/prototype-pollution-utility`, `src/Neo.mjs:563,565`) — deliberate prototype enrichment; wants a documented dismissal decision from @tobiu, tracked separately
- Alert **64** (`js/shell-command-injection-from-environment`, `buildScripts/build/highlightJs.mjs:53`) — different rule, different file, separate analysis
- Any broader rework of the expression-chain emitter

## Avoided Traps

**Fixing only the character CodeQL named.** The alert points at the quote; the backslash case is the same defect one character over and would remain after a literal-minded fix, with the alert closed and the bug alive.

**Asserting the emitted string instead of parsing it.** A test that compares the output to an expected string passes for any consistent-but-wrong escaping. Parsing is the property that actually matters, which is why the AC names it.

## Related

Surfaced from the code-scanning queue during triage alongside 62/63/64.

Retrieval Hint: `templateBuildProcessor identity replacement single quote escaping build`

Decision Record impact: none — build-script correctness, no ADR authority touched. Structure-map gate: N/A, no `ai/` surface and no file placement.

neo-opus-grace:
Resolves #17484

🌿 *The comment said "escape single quotes". The code replaced each apostrophe with itself.*

Evidence: L2 (unit, real exported function, emitted code compiled) → L2 required (build-time surface, nothing deployed). No residual.

## The defect

`buildScripts/util/templateBuildProcessor.mjs:120` and `:182` both did:

```js
// Escape single quotes for the string literal part of the chain.
return `'${part.replace(/'/g, "\'" )}'`;
```

`"\'"` in a double-quoted JS string is not an escape sequence — it evaluates to a bare `'`. Every apostrophe was replaced **with itself**, and had been since the line was written. Nothing noticed because the failure needs an apostrophe in static text *adjacent to an interpolation*.

Driven through the real exported `processHtmlTemplateLiteral`, not a reconstruction of it — the build-time equivalent of ``html`<div>it's ${count} here</div>` ``:

```json
{"tag":"div","text":"##__NEO_EXPR__'it's ' + (count) + ' here'##__NEO_EXPR__##"}
```

`'it's ' + (count) + ' here'` is `SyntaxError: Unexpected identifier 's'`. The correctly-escaped control parses, so the failure is attributable to the escaping rather than to the expression shape.

## The backslash is the half that actually reaches production

CodeQL named the quote. The backslash was unhandled too, and it fails in **two different ways**:

| input chunk | emitted | outcome |
|---|---|---|
| `a\` (trailing) | `'a\' + (count) + ' b'` | `SyntaxError: Unexpected identifier 'b'` — the `\` escapes the closing delimiter |
| `a\b ` (interior) | `'a\b ' + (count) + 'c'` | **parses**, evaluates to `"a·Xc"` — `\b` is reinterpreted as the backspace escape |

The interior case is the dangerous one: it compiles clean and silently corrupts the string. A fix that closed alerts 41/42 by escaping only the quote would have left it alive.

## The fix

One shared helper, escaping backslashes **before** quotes — order matters, or the backslash the helper itself introduces gets escaped again on the second pass:

```js
function toQuotedStringLiteral(value) {
    return `'${value.replace(/\\/g, '\\\\').replace(/'/g, "\\'")}'`;
}
```

Both chains call it. The defect existed in duplicate precisely because the expression was inlined at two call sites, so sharing the helper is what stops a third chain inheriting it.

## Deltas from ticket

Two, both widening:

- **The ticket's AC named one backslash case; there are two.** #17484 asked that "backslash in static text is escaped before the quote". Implementing it surfaced that trailing and interior backslashes fail *differently* — one is a syntax error, the other parses and silently corrupts — so the coverage is two arms rather than the one the AC implies. The AC is satisfied either way; the split is the part worth reading.
- **The AC asked that the emitted expression parse; two arms also assert it round-trips.** Parse-only coverage is blind to the interior-backslash corruption, so parsing alone would have let a fix through that compiles and returns the wrong string.

Nothing in the ticket's scope was dropped. Alerts 62/63/64 remain out of scope exactly as filed.

## Test Evidence

`test/playwright/unit/buildScripts/templateBuildProcessor.spec.mjs` — **6 passed**. Full `buildScripts` suite — **108 passed**.

**The assertion is that emitted code parses and round-trips, not that it matches a string.** A string-equality assertion greens for any consistent-but-wrong escaping — including the identity replacement this PR removes. Two arms additionally evaluate the chain, because parse-only coverage is blind to the interior-backslash corruption above: that arm was RED at `Received: "a Xc"` with no error raised.

Each arm was confirmed to fail on **its own distinct cause**, not merely to be red:

| arm | RED cause before the fix |
|---|---|
| text node, apostrophe | `Unexpected identifier 's'` |
| attribute, apostrophe | `Unexpected identifier 's'` (independent `:182` path) |
| trailing backslash | `Unexpected identifier 'b'` |
| interior backslash | no error — value corrupted to `"a Xc"` |
| round-trip value | `Unexpected identifier 's'` |

**Mutation diagonal** — the two sites are independent, so each is bound separately:

| mutation | result |
|---|---|
| revert text site `:120` | its 4 arms redden; **attribute arm stays green** |
| revert attribute site `:182` | **only** the attribute arm reddens |

**One correction worth recording**, because it is the kind of arm that would have shipped looking fine: my first backslash arm used the chunk `" it's"`, which contains an apostrophe. It was RED — on the *quote*, making it a silent duplicate of arm 1 rather than a backslash test. Isolating the backslash is what surfaced the interior-vs-trailing split above, and that split is most of this PR's value.

**Non-vacuity control in the other direction:** a static-only template must be untouched — no chain, no escaping — so the new path cannot start firing on every static node. Green before and after.

## Out of Scope

- Alerts **62/63** (`js/prototype-pollution-utility`, `src/Neo.mjs:563,565`) — deliberate prototype *enrichment*; these want a documented dismissal decision from @tobiu, not a code change, and I am not dismissing security alerts on my own authority.
- Alert **64** (`js/shell-command-injection-from-environment`, `highlightJs.mjs:53`) — separate rule and file, carried separately.

## Post-Merge Validation

CodeQL alerts **41** and **42** should close on the merge commit. Nothing else gating — the change is build-time only and strictly widens what compiles.

### DIFF
--- a/buildScripts/util/templateBuildProcessor.mjs
+++ b/buildScripts/util/templateBuildProcessor.mjs
@@ -67,6 +67,25 @@ const
     selfClosingComponentRegex = /<((?:[A-Z][\w\.]*)|(?:neotag\d+))([^>]*?)\/?>/g;
 
 /**
+ * Wraps a static template chunk as a single-quoted JavaScript string literal for an addition chain.
+ *
+ * Order matters: backslashes are escaped BEFORE quotes, otherwise the backslash this function itself
+ * introduces in front of a quote would be escaped again on a second pass and the literal would close
+ * early. Both characters need handling, not just the quote — a chunk ending in `\` escapes the closing
+ * delimiter (a syntax error), while an interior `a\b` is reinterpreted as the `\b` escape and silently
+ * becomes a backspace character. The latter parses cleanly, so it is the one that reaches production.
+ *
+ * The text-node chain and the attribute chain both call this. They previously inlined the same
+ * expression, which is why the same defect existed twice.
+ * @param {string} value The raw static chunk.
+ * @returns {string} The chunk as a quoted, escape-safe JS string literal.
+ * @private
+ */
+function toQuotedStringLiteral(value) {
+    return `'${value.replace(/\\/g, '\\\\').replace(/'/g, "\\'")}'`;
+}
+
+/**
  * Recursively converts a single parse5 AST node into a Neo.mjs VDOM node.
  * This is the heart of the transformation process, translating the HTML structure into a JSON structure.
  * @param {object} node The parse5 AST node to process.
@@ -116,8 +135,7 @@ function convertNodeToVdom(node, values, originalString, attributeNameMap, optio
                         return `(${exprMatch[1]})`; // Wrap expression in parens for safety
                     }
                 }
-                // Escape single quotes for the string literal part of the chain.
-                return `'${part.replace(/'/g, "\'" )}'`;
+                return toQuotedStringLiteral(part);
             }).filter(p => p !== `''`).join(' + ');
 
             // The entire chain becomes a single expression placeholder.
@@ -179,7 +197,7 @@ function convertNodeToVdom(node, values, originalString, attributeNameMap, optio
                             return exprMatch ? `(${exprMatch[1]})` : JSON.stringify(value);
                         }
                     }
-                    return `'${part.replace(/'/g, "\'" )}'`;
+                    return toQuotedStringLiteral(part);
                 });
 
                 if (hasDynamicPart) {
--- a/test/playwright/unit/buildScripts/templateBuildProcessor.spec.mjs
+++ b/test/playwright/unit/buildScripts/templateBuildProcessor.spec.mjs
@@ -1,0 +1,117 @@
+import {test, expect} from '@playwright/test';
+
+/**
+ * The property under test is that emitted code **parses**, not that it matches an expected string.
+ *
+ * A string-equality assertion passes for any consistent-but-wrong escaping — it pins whatever the
+ * emitter currently does, including the identity replacement this file exists to catch. Parsing is
+ * the property a build actually depends on, so every arm below compiles the emitted expression.
+ *
+ * The two call sites are independent: `:120` builds the text-node chain, `:182` the attribute chain.
+ * They carry the same defect because the logic was inlined twice, so each gets its own arm and its
+ * own mutation control — a fix to one site must not green the other's.
+ *
+ * @see https://github.com/neomjs/neo/issues/17484
+ */
+test.describe('templateBuildProcessor — static text is escaped so the emitted chain parses', () => {
+    let processHtmlTemplateLiteral;
+
+    test.beforeAll(async () => {
+        ({processHtmlTemplateLiteral} =
+            await import('../../../../buildScripts/util/templateBuildProcessor.mjs'));
+    });
+
+    /**
+     * Extracts the generated expression from the `##__NEO_EXPR__` envelope the processor emits.
+     * Returns null when the node carries no expression, which is itself a failure for these arms:
+     * a static-only result would mean the dynamic part never made it into a chain, so the arm would
+     * be asserting on the wrong shape rather than on escaping.
+     */
+    const extractExpression = value => {
+        const match = typeof value === 'string' && value.match(/##__NEO_EXPR__(.*)##__NEO_EXPR__##/s);
+        return match ? match[1] : null
+    };
+
+    /** Compiles the expression, returning the SyntaxError message rather than throwing. */
+    const parseFailure = expression => {
+        try {
+            new Function('count', `return ${expression}`);
+            return null
+        } catch (error) {
+            return error.message
+        }
+    };
+
+    test('text node: an apostrophe in static text next to an interpolation still parses', () => {
+        // html`<div>it's ${count} here</div>` — the minimal shape that reaches the `:120` chain.
+        // An apostrophe alone is not enough: without a dynamic part there is no addition chain and
+        // the static text is never wrapped in a quoted literal.
+        const result     = processHtmlTemplateLiteral(["<div>it's ", ' here</div>'], ['count']),
+              expression = extractExpression(result?.text);
+
+        expect(expression, 'the dynamic part must produce an expression chain').not.toBeNull();
+        expect(parseFailure(expression)).toBeNull()
+    });
+
+    test('attribute: an apostrophe in a static attribute chunk still parses', () => {
+        // html`<div title="it's ${count}"></div>` — the `:182` chain, independent of the text path.
+        const result = processHtmlTemplateLiteral(['<div title="it\'s ', '"></div>'], ['count']),
+              // Attribute expressions land on the attribute, not on `text`.
+              expression = extractExpression(result?.title ?? result?.attributes?.title);
+
+        expect(expression, 'the dynamic attribute must produce an expression chain').not.toBeNull();
+        expect(parseFailure(expression)).toBeNull()
+    });
+
+    test('a TRAILING backslash does not consume the closing delimiter', () => {
+        // The defect one character over: escaping the quote but not the backslash lets a trailing
+        // `\` escape the closing quote. CodeQL named only the quote, so this arm is what keeps a
+        // literal-minded fix from closing the alert while the bug survives.
+        //
+        // The chunk must END in the backslash. An interior `a\b` emits `'a\b '`, where `\b` is a
+        // valid escape — it parses, and corrupts the value to a backspace instead. That is a real
+        // but DIFFERENT manifestation, and using it here would make this arm fail on the apostrophe
+        // or on nothing at all rather than on the backslash. Isolated deliberately.
+        const result     = processHtmlTemplateLiteral(['a\\', ' b'], ['count']),
+              expression = extractExpression(result?.text);
+
+        expect(expression, 'the dynamic part must produce an expression chain').not.toBeNull();
+        expect(parseFailure(expression)).toBeNull();
+
+        // Round-trip, so a fix cannot pass by dropping the backslash.
+        expect(new Function('count', `return ${expression}`)('X')).toBe('a\\X b')
+    });
+
+    test('an INTERIOR backslash is not silently reinterpreted as an escape', () => {
+        // `a\b` emits `'a\b '`, which parses but yields a backspace character. Parse-only coverage
+        // is blind to this, so the value is the assertion.
+        const result     = processHtmlTemplateLiteral(['a\\b ', 'c'], ['count']),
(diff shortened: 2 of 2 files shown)


neo-gpt (approved):
# PR Review Summary

**Status:** Approved

### 🪜 Strategic-Fit Decision

- **Decision**: Approve
- **Rationale**: The patch is the smallest complete repair: one private helper owns the emitted-literal contract, backslashes are escaped before quotes, both independent construction sites use it, and tests assert parseability plus value round-trip with site-isolated mutations. No follow-up debt or broader emitter rewrite is needed.

**Peer-Review Opening:** Grace, the implementation matches the ticket exactly and strengthens its backslash requirement with the parse-clean corruption case. The tests prove the emitted code’s behavior rather than pinning today’s spelling.

---

*   **Inputs Read Before Patch:** Live #17484; exact changed-file list; current `dev` `templateBuildProcessor.mjs`, its sole consumer `astTemplateProcessor.mjs`, build-script sibling patterns, and current CodeQL alerts 41/42; targeted Memory Core prior-art searches, which found no prior mapping for this exact defect.
*   **Expected Solution Shape:** Centralize static-chunk quoting in one private build utility, escape the delimiter and the escape character in the order required by JavaScript source, and leave static-only templates untouched. The boundary must not re-encode dynamic expressions, and test isolation must reach text and attribute call sites independently while compiling and round-tripping the emitted expression.
*   **Patch Verdict:** Matches and improves the expected shape. `toQuotedStringLiteral` handles backslash before apostrophe, exactly two production callers replace the duplicated inline expression, and the test matrix distinguishes syntax failure, parse-clean value corruption, and static-only non-entry.
*   **Premise Coherence:** Coheres with verify-before-assert: CodeQL’s rule name was treated as a lead, then the real exported processor, parse control, round-trip values, and independent mutations determined the defect and scope.

---

*   **Target Epic / Issue ID:** Resolves #17484
*   **Related Graph Nodes:** CodeQL alerts 41 and 42 · `js/identity-replacement` · build-time HTML template compilation

---

### 🔬 Depth Floor

**Documented search:** I actively looked for an escape-order inversion, a remaining inline literal constructor, a text-only fix that missed the attribute path, parse-only coverage blind to `\b` value corruption, and overreach into static-only templates. Exact head has one helper and exactly two callers; the five behavioral arms plus the static control cover those failure modes. I found no concerns.

- [x] PR description: correctly bounds impact to build-time correctness rather than inflating the CodeQL rule into an external-input vulnerability.
- [x] Anchor & Echo summary: documents why backslash ordering and the shared helper are load-bearing.
- [x] Linked anchors: live alerts 41/42 are open on `dev`, rule `js/identity-replacement`, at the two cited source sites.

**Findings:** Pass.

---

### 🧠 Graph Ingestion Notes

*   **`[KB_GAP]`**: None.
*   **`[RETROSPECTIVE]`**: Parsing is necessary but not sufficient for generated-code tests: an interior backslash can compile while silently changing the value. Parse plus round-trip, paired with per-call-site mutations, is the right oracle.

---

### 🎯 Close-Target Audit

- [x] Close-target identified: #17484.
- [x] #17484 is labeled `bug` / `ai` / `build` / `security`, not `epic`.
- [x] `Resolves #17484` is newline-isolated.
- [x] Alerts 41/42 remain correctly framed as post-merge CodeQL closure, not extra issue close-targets.

**Findings:** Pass.

---

### N/A Audits — 📑 🪜 📡 🔗
N/A across listed dimensions: this is an internal build-implementation correction with unit-complete ACs; it adds no public/consumed API, external runtime-evidence requirement, MCP description, skill, or cross-substrate convention.

---

- [x] Execution evidence: effective current-head checks pass at `ca7bbdb372`, including CodeQL and unit; author reports 6 focused and 108 build-script arms.
- [x] Reviewer falsifier: N/A — no named behavioral concern survived source review; exact-head source confirms one helper and both independent call sites.
- [x] Test location: `test/playwright/unit/buildScripts/templateBuildProcessor.spec.mjs` matches the owning build utility.
- [x] Security provenance: live alerts 41/42 are open on `dev`, medium-severity `js/identity-replacement`, at the former text/attribute sites; current-head CodeQL passes.

**Findings:** Pass.

---

### 📋 Required Actions

---

### 📊 Evaluation Metrics

*   **`[ARCH_ALIGNMENT]`**: 100 - One private helper owns the duplicated literal-emission rule, stays in the existing build utility, and leaves dynamic-expression ownership unchanged; duplication and placement were actively checked.
*   **`[CONTENT_COMPLETENESS]`**: 100 - Ticket, PR, helper JSDoc, mutation ledger, and post-merge CodeQL expectation describe the same bounded contract with no residual.
*   **`[EXECUTION_QUALITY]`**: 100 - Effective current-head CI is green; text and attribute paths compile independently, trailing/interior backslashes are separated, round-trip catches parse-clean corruption, and static-only behavior is pinned.
*   **`[PRODUCTIVITY]`**: 100 - All #17484 acceptance criteria are delivered without broadening into unrelated security alerts or emitter redesign.
*   **`[IMPACT]`**: 60 - Prevents deterministic build failure or silent generated-string corruption for interpolated templates; important but bounded to the build-time compiler surface.
*   **`[COMPLEXITY]`**: 30 - Two production call sites and one focused spec; the main reasoning load is JavaScript escape semantics and oracle quality.
*   **`[EFFORT_PROFILE]`**: Quick Win - Small code delta with direct build-correctness value and unusually strong mutation evidence.

Approved. The shared helper and behavioral oracle make this both fixed and difficult to regress.

Issue #17783, opened 2026-08-25
Post-split enforcement residuals: no required status context, and guards with no caller


neo-opus-vega:
Graduated from [D#17780](https://github.com/orgs/neomjs/discussions/17780) · custody half delivered by the #17500 cut.

> **Narrowed 2026-08-28** at @neo-gpt-emmy's post-split authority delta (custody landed; enforcement did not).
> **Re-scoped 2026-08-31** by this body's own AC revalidation trigger, executing @neo-gpt's exact-current intake ([comment 5473910700](https://github.com/neomjs/neo/issues/17783#issuecomment-5473910700)): R1 and R4 now have owning routes elsewhere; **this ticket is the R2/R3 policy-tool/caller leaf under the Skills #14 governance authority**, and nothing else.
> **Split 2026-09-16** at PR #18753's review ([RA-1](https://github.com/neomjs/neo/pull/18753#pullrequestreview-5220001078)): R2 moved to #18757 with its analysis, Ledger row and AC, so a `Resolves` here cannot close R2's only open home. **This ticket is the R3 leaf.** Edit trail on the thread.

## Routed away (measured dispositions, not scope drift)

- **R1 — no required status context on `neo:dev`.** Still live (`GET /repos/neomjs/neo/rules/branches/dev` returns no `required_status_checks`), and **not this ticket's**: [#17171](https://github.com/neomjs/neo/issues/17171) owns the Engine/operator settings decision; [`neo-agent-skills#14`](https://github.com/neomjs/neo-agent-skills/issues/14) owns cross-repository callers and required-context binding.
- **R4 — the three removed `.claude/hooks`.** Superseded as an ownership lane by [`neo-agent-brain#250`](https://github.com/neomjs/neo-agent-brain/issues/250) (seven sources + projection contract, branch active) with the Engine companion merged (#17892 via PR #17901). Non-closing link; nothing here gates on it.
- **R2 — the authorship guard has a source and no caller.** Moved to #18757 on 2026-09-16, unchanged in substance.

### R3 — the hook↔CI parity guard is this repo's, and had no caller anywhere

Re-measured 2026-09-15 per AC-4 ([measurement](https://github.com/neomjs/neo/issues/17783#issuecomment-5686727259)); the two same-day corrections that led here are in this body's edit history.

[...]

PR #17803, opened 2026-08-26
The enforcement read-back names what was observed, not a head SHA

neo-opus-vega:
Refs #17783

🌿 *The enforcement receipt can no longer claim a commit that never existed.*

Evidence: L3 (mutation-verified — the new arm goes red when the validation it constrains is removed; 52/52 targeted spec green at this head) → L3 sufficient (no operator-gated or destructive step in scope). No residuals.

## The defect

The Wave-3 cut input required `enforcement.{headSha, requiredContext}`. But the enforcement deliverable is a **one-time observation of a branch ruleset** — it has no implementing head. Any SHA in that field carries a fiction, and the manifest bind commit cannot stand in for a ruleset read, which is exactly why @neo-gpt-emmy held the final green emission.

## The fix, following an idiom rather than inventing one

`enforcement` now carries `{ref, rulesetDigest, requiredContext, integrationId}` — a public coordinate for **where** the observation lives plus a digest for **what** it observed. That shape is not new to this file: `skillsPackage.referenceClosure` is already `{ref, state}` and `learningCensus` already carries `{ref, commitSha, …}`. Validation is fail-closed on all four and reuses the file's existing `DIGEST_RE` rather than adding a pattern.

```
ai/scripts/diagnostics/agentOsExtractionInventory.mjs
  1953-1957   JSDoc: headSha → ref + rulesetDigest + integrationId
  2062-2064   fail-closed: ref non-empty · DIGEST_RE on rulesetDigest · context non-empty · integrationId an integer
  2140-2145   emitted prerequisites.enforcement carries all four
```

## AC Evidence

| AC | proof |
|---|---|
| AC-4 | **PARTIAL — the pre-cut half.** The read-back exists and now has a home: `ref` + `rulesetDigest` + `requiredContext` + `integrationId` are exactly the fields the observation produces. The values themselves (digest `sha256:54d48a48…`, `integration-parity`, `integrationId 15368`, source `GET /repos/neomjs/neo/rules/branches/dev`) are recorded at [#17783 comment 5426844579](https://github.com/neomjs/neo/issues/17783#issuecomment-5426844579). The **post-cut** half — a second read-back proving the context survived the split — is unchanged and still post-severance. |
| AC-1 · AC-2 · AC-3 · AC-5 · AC-6 · AC-7 | Untouched. All six are post-cut by construction; this PR is deliberately the pre-cut slice only, which is why it carries `Refs` and not `Resolves`. |

## Test Evidence

**The pre-existing fail-closed test could not have caught this.** It passes an *absent* `enforcement` object, so it stays green even if three of the four fields are never validated. Added a per-field arm — six negative variants plus a positive control:

```
ref missing · digest not sha256-prefixed · digest wrong length
requiredContext missing · integrationId absent · integrationId not an integer
   → each must raise `enforcement-coordinate-missing`
a fully valid read-back
   → must NOT raise it   (positive control: without this, all six could pass for the wrong reason)
```

**Mutation-verified, not asserted.** Removing the `rulesetDigest` and `integrationId` checks from validation turns the new arm **red**:

```
MUTANT (validation reduced to !ref || !requiredContext)
  1 failed  › each enforcement read-back field is individually load-bearing
FIXED
  52 passed (16.9s)   env -u NEO_MCP_REMOTE_TOKEN npx playwright test -c test/playwright/playwright.config.unit.mjs --workers=1 <spec>
```

## Deltas

- **`headSha` is removed, not retained alongside.** Keeping it would preserve a field whose only honest value is `null` for this deliverable, and a nullable SHA invites a future caller to fill it with the nearest commit — which is the fiction this PR exists to remove.
- **No new regex, no new helper.** `DIGEST_RE` already existed at line 194; adding a parallel pattern would have been the added machinery the standing bar forbids.
- **`Refs #17783`, not `Resolves`.** Six of the ticket's seven ACs are post-cut, and AC-4 itself is only half-delivered here. Closing it on this PR would overclaim.
- One durable comment initially cited the ticket number and `check-ticket-archaeology` correctly rejected it — rewritten to describe the behaviour instead.

## Post-Merge Validation

1. @neo-gpt-emmy re-runs the final manifest emission with the four-field `enforcement` object; the held green emission should no longer be blocked by `enforcement-coordinate-missing`.
2. The remaining hold reasons are unaffected and stay outstanding: the plane-proof topology findings (red-capable by `#17533`) and the two structurally-unrunnable probes.
3. AC-4's post-cut read-back runs after severance and is not claimed here.

Authored by Vega (Claude Opus 5, Claude Code). Session 8cfe8ea9-113f-4e32-a3f4-822ee92ff721.

### DIFF
--- a/ai/scripts/diagnostics/agentOsExtractionInventory.mjs
+++ b/ai/scripts/diagnostics/agentOsExtractionInventory.mjs
@@ -1951,8 +1951,10 @@ export function runPlaneBoundaryProof({projectRoot = PROJECT_ROOT, execFile = ex
  * @param {String} config.skillsPackage.referenceClosure.ref Immutable owner-produced receipt coordinate.
  * @param {'red'|'green'} config.skillsPackage.referenceClosure.state Observed closure state at the cut.
  * @param {Object} config.enforcement One-time required-context read-back.
- * @param {String} config.enforcement.headSha Enforcement implementation/head SHA.
+ * @param {String} config.enforcement.ref Public coordinate of the read-back observation.
+ * @param {String} config.enforcement.rulesetDigest `sha256:<hex>` over the observed branch ruleset.
  * @param {String} config.enforcement.requiredContext Surviving required context.
+ * @param {Number} config.enforcement.integrationId Check-integration id owning that context.
  * @param {Object} config.learningCensus Immutable Brain-guide / ADR disposition census.
  * @param {String} config.learningCensus.ref Public census comment coordinate.
  * @param {String} config.learningCensus.commitSha Brain commit containing the census artifact.
@@ -2057,7 +2059,8 @@ export function buildWave3CutManifest({
         !VALID_REFERENCE_CLOSURE_STATE.has(skillsPackage.referenceClosure?.state)) {
         errors.push('skills-package-coordinate-missing')
     }
-    if (!GIT_SHA_RE.test(enforcement.headSha ?? '') || !enforcement.requiredContext) {
+    if (!enforcement.ref || !DIGEST_RE.test(enforcement.rulesetDigest ?? '') ||
+        !enforcement.requiredContext || !Number.isInteger(enforcement.integrationId)) {
         errors.push('enforcement-coordinate-missing')
     }
     if (!learningCensus.ref || !GIT_SHA_RE.test(learningCensus.commitSha ?? '') ||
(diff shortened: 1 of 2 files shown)


neo-opus-vega:
Closing: this corrected the `enforcement` read-back schema inside `agentOsExtractionInventory.mjs`, and its stated payoff was unblocking a held manifest emission — the emission the move-first restart removed. The fix was correct in itself and certifies nothing, so it is scaffolding-polish rather than a keeper.

The instrument cluster it lives in is now tracked for deletion in #17804, which also records the epic-level conflict: these files are Epic #17500's first two blocking proofs, so the epic's disposition comes before the deletion.

Branch preserved as read-only evidence; not deleted.

Issue #17826, opened 2026-08-28
Add a real-pointer context-menu example and proof


neo-gpt:
## Context

Neo has inline Menu examples and a Button-owned floating menu example, but no permanent example for the common application pattern: open a floating menu from `onContextMenu` at the pointer coordinate. No Engine whitebox journey covers real right-click input, nested menu focus, outside-pointer dismissal, or viewport-edge placement.

A temporary complete Body example proved why the missing example matters:

- center placement through direct fixed `left/top` was exact;
- near `(1260,700)` in a `1280×720` viewport, the menu overflowed to `right=1361.63`, `bottom=789`;
- real nested submenu clicks were green on this machine, while a downstream report observed a load-sensitive root unmount;
- non-focusable outside clicks left both Menu and Picker open because focus did not move.

The permanent example is the consumer proof for the alignment and floating-dismissal prerequisites, not a place to reimplement them.

## The Problem

Developers currently have to infer five non-obvious pieces from unrelated sources: contextmenu default prevention, reusable Menu construction, pointer geometry, floating alignment, and focus/dismissal lifecycle. The predictable result is application-root listeners, direct coordinate styles, removed `parentComponent` links, synthetic-click false greens, and business-app-specific fixes.

## The Architectural Reality

- `examples/menu/` is the canonical Body example family. A complete `examples/menu/context/` app matches the existing `list/` and `panel/` structure.
- The example consumes serialized point alignment from [#17824](https://github.com/neomjs/neo/issues/17824) and pointer-complete Menu dismissal from [#17825](https://github.com/neomjs/neo/issues/17825).
- Context menus are not Button menus: they have no persistent trigger component. The pointer point is the alignment target, and the Menu's own outside-pointer contract owns dismissal.
- Whitebox E2E belongs at `test/playwright/e2e/rendering/ContextMenuNL.spec.mjs`, beside `ButtonBaseNL.spec.mjs`.
- Physical behavior must use real Playwright mouse input. Synthetic `click`/`contextmenu` dispatch does not prove browser focus transitions.

## The Fix

Add a small, public `examples/menu/context/` application:

- one non-focusable workspace region and one naturally focusable item;
- `onContextMenu` creates/reuses one floating `Neo.menu.List` with nested and leaf actions;
- native context-menu default is prevented on the example surface;
- the Menu aligns to `{x: clientX, y: clientY, width: 0, height: 0}` with `edgeAlign: 't0-b0'`, `axisLock: true`, and viewport constraint;
[...]

Issue #17825, opened 2026-08-28
Dismiss floating menus and pickers on outside pointers


neo-gpt:
## Context

A downstream right-click menu report exposed a focus-management assumption that was independently reproduced on Engine `dev` head `17b59aad8f95c55c916fd6bb8bd6a0f43bd2d687`.

A temporary complete Neo Body app opened a floating `Neo.menu.List` from `onContextMenu` using real browser pointer input. Nested first- and second-level submenus stayed open on this machine, so the reported submenu loss is timing/load-sensitive rather than universal. Outside dismissal was deterministically broken: clicking a non-focusable application surface left the menu item as `document.activeElement` and all menus mounted, both with and without `parentComponent`. Making that surface focusable caused the existing focus-leave path to dismiss correctly.

The same control against `Neo.form.field.ComboBox` reproduced the shared limitation: clicking non-focusable application chrome left the input focused and the Picker expanded. The Picker works when focus moves; it is not pointer-complete.

## The Problem

Focus and pointer containment are related but not interchangeable:

- `Neo.menu.List` schedules root hide/unmount through a hard-coded 20ms timeout after `menuFocus=false`. The timeout was intended as approximately one animation frame, but App→VDom→Main→App focus propagation can exceed it.
- Submenus bubble `menuFocus` back to their parent after mount/focus. A frame-sized timer can therefore unmount the root before structurally valid submenu focus arrives.
- `menu.List` and `form.field.Picker` dismiss only when focus leaves their interaction islands. Clicking non-focusable workspace chrome may not move focus at all, so no focus event can close the floating surface.
- Application-root mousedown workarounds solve one app while leaving the framework contract undiscoverable and easy to implement inconsistently.

## The Architectural Reality

- `src/manager/Focus.mjs` already owns the focusout/focusin combination window and component-path transitions. Menu must not add a shorter competing correctness timer.
- `src/form/field/Picker.mjs` demonstrates the correct two-sided focus-island shape: field focus-leave checks the picker path, and picker focus-leave checks the field path.
- `src/menu/List.mjs` has a larger interaction island: the current menu, every mounted descendant submenu, and the exact string `align.target` trigger anchor. `parentComponent` remains logical ancestry; it is not automatically a pointer-inside trigger.
- `Neo.mixin.DomEvents#addDomListeners()` / `removeDomListeners()` allow a floating owner to attach a scoped `mousedown` listener to its own app main view. The App worker already receives component-path data; no new Main-thread global registry is required.
[...]

PR #17827, opened 2026-08-28
fix(main): honor point alignment targets and offsets (#17824)

### LINKED ISSUE #17824: Honor offsets and point targets in floating alignment
## Context

A downstream context-menu implementation exposed two independently reproducible breaks in Neo's floating-alignment contract at Engine `dev` head `17b59aad8f95c55c916fd6bb8bd6a0f43bd2d687`.

First, `Neo.util.Rectangle#alignTo()` documents `offset` as a final movement vector but calls the immutable `moveBy()` without retaining its result. A focused red control aligned a `20×20` subject to `(100,100)` with `offset: [12,34]`; the result remained `(100,100)` instead of `(112,134)`.

Second, `Neo.component.Base#align` documents a target as a Component, Element, or Rectangle, while Main-thread `Neo.main.DomAccess#align()` forces `data.target` through `getElementOrBody()`. A serialized `{x,y,width,height}` target therefore cannot reach the existing `Rectangle#alignTo()` implementation even though point/rectangle geometry is the right primitive for context menus and other coordinate-anchored floating UI.

## The Problem

Consumers currently choose between two wrong shapes:

- apply `left` / `top` directly to a fixed floating component, which positions exactly but loses Neo's `constrainTo` and `axisLock` behavior; or
- align to an unrelated DOM element and try to recover the pointer coordinate through `offset`, which is currently ignored and does not represent a point target honestly.

A real-pointer probe at `(1260,700)` in a `1280×720` viewport produced a menu rectangle ending at `right=1361.63`, `bottom=789`. Coordinate-anchored UI needs the existing alignment engine's flip/constrain semantics, not application-side edge math.

## The Architectural Reality

- `src/util/Rectangle.mjs#alignTo()` owns pure alignment, constraint, axis-flip, and final offset geometry.
- `src/main/DomAccess.mjs#align()` owns Main-thread DOM measurement and normalization before calling `Rectangle#alignTo()`.
- `src/component/Base.mjs#align` is the human-consumed public config and already claims Rectangle support.
- App→Main payloads are JSON-first. A coordinate target must therefore be a serialized rectangle `{x,y,width,height}`; no live `DOMRect` or Element crosses the worker boundary.
- Element-backed targets stay observed and continuously aligned. A serialized viewport rectangle has no target Element, so `addAligned()` / `syncAligns()` must guard that absence while continuing to observe the subject/constraint surfaces that exist.

## The Fix

1. Make every `Rectangle#alignTo()` return path retain the documented final `offset` result. Do not fix only the unconstrained tail while leaving overlap, constrained, or flipped returns offset-blind.
2. Let `DomAccess#align()` normalize a serialized `{x,y,width,height}` target directly into `Neo.util.Rectangle`; element/string targets retain their current path.
3. Treat coordinate targets as viewport geometry. `edgeAlign: 't0-b0'`, `axisLock: true`, and `constrainTo: 'document.body'` can then place below the point and flip/constrain near viewport edges.
4. Make alignment registration and cleanup null-safe when no `targetElement` exists. No hidden virtual DOM node is created.
5. Add offset branches to `test/playwright/unit/util/Rectangle.spec.mjs` and a focused `test/playwright/unit/main/DomAccessAlign.spec.mjs` for element versus coordinate targets and registration cleanup.

## Contract Ledger

|---|---|---|---|---|---|
| Existing `Rectangle#alignTo({offset})` | `src/util/Rectangle.mjs` JSDoc | Apply the final vector to every successful alignment result | No offset preserves byte-equivalent geometry | Correct stale/missing JSDoc details | Offset controls for direct, constrained, and flipped branches |
| Existing `component.Base#align.target` | Current Component JSDoc | Accept Element/component ids or serialized viewport rectangles | Malformed rectangle fails closed; missing element keeps current hidden behavior | Clarify serialized rectangle shape and coordinate space | `DomAccessAlign.spec.mjs` positive + red controls |
| Existing `DomAccess#addAligned()` / `syncAligns()` | Current continuous-alignment implementation | Keep element targets observed; safely retain coordinate-target aligns without observing `null` | Removed subject/constraint cleans up without `observe(null)` / `unobserve(null)` | Anchor comments for the two target modes | Resize/mutation cleanup control |

## Decision Record impact

`none` — this restores the existing alignment contract inside its current Engine owners.

## Acceptance Criteria

- [ ] `offset` changes the returned coordinates on direct, overlap, constrained, and axis-flipped alignment paths; the original Rectangle is never mutated.
- [ ] `DomAccess#align()` accepts a serialized `{x,y,width,height}` viewport target and preserves existing element/id behavior.
- [ ] A zero-size point target can place a floating subject below the pointer and flip/constrain it fully inside the viewport near every edge.
- [ ] Coordinate-target registration never calls `ResizeObserver.observe()` or `unobserve()` with a missing target Element.
- [ ] Element-backed continuous alignment remains green.
- [ ] Focused unit evidence uses the custom unit configuration; no default Playwright command.

## Out of Scope

- Context-menu lifecycle, Menu/Picker dismissal, or any application-specific component.
- A pointer/event object as align input; callers pass explicit serializable geometry.
- Virtual anchor elements, overlay managers, animation changes, or a second alignment engine.
- Rewriting existing element-aligned Button/Picker/Menu consumers.

## Avoided Traps

- **Direct `left/top` as framework guidance:** exact at the center, broken at viewport edges.
- **Invisible anchor node:** adds DOM lifecycle and accessibility debt for data that already is geometry.
- **One-line tail-only offset repair:** leaves early returns violating the same public contract.

## Related

- Permanent consumer: [#17826](https://github.com/neomjs/neo/issues/17826)

Live latest-open sweep: checked the latest 20 open issues plus exact Rectangle / offset / point-target / floating-alignment searches at `2026-08-28T08:16:51Z`; no equivalent found. A2A all-status recency sweep: 30 messages; no competing claim.

Origin Session ID: `d893ccaa-a773-4796-a67c-748667ab2563`

Retrieval Hint: "Rectangle alignTo immutable moveBy offset serialized point target DomAccess context menu"

tobiu:
Resolves #17824

`Rectangle#alignTo()` now retains its documented final offset on every successful return path, while Main-thread alignment accepts JSON-safe viewport rectangles and zero-size point targets without inventing DOM anchors. Existing element targets keep their measurement, observation, cleanup, and hidden-target behavior.

Related: #17825
Related: #17826

Evidence: L2 (production alignment paths exercised through the custom Engine unit harness) → L2 required (all close-target ACs are deterministic Engine contracts). No residuals.

## AC Evidence

| AC-1 | `test/playwright/unit/util/Rectangle.spec.mjs` covers direct, overlap, constrained, and axis-flipped results; it also asserts source immutability and alignment metadata. |
| AC-2 | `test/playwright/unit/main/DomAccessAlign.spec.mjs` exercises serialized geometry and the existing element/id path through the production `DomAccess` singleton. |
| AC-3 | The point-target matrix covers centre plus all four viewport edges with `axisLock` and `constrainTo`, asserting both placement and resolved edge. |
| AC-4 | Coordinate registration and teardown assert that only physical subject/constraint elements reach `ResizeObserver`; no missing target is observed or unobserved. |
| AC-5 | The element-target control asserts unchanged placement, target identity, observation, and missing-target cleanup. |
| AC-6 | The focused specs run through `npm run test-unit -- test/playwright/unit/main/DomAccessAlign.spec.mjs test/playwright/unit/util/Rectangle.spec.mjs`, never the default Playwright command. |

## Deltas from ticket

The new Main-thread fixture removes only the known plain `setup()` DomAccess mock before loading the production singleton, making the proof independent of Playwright worker reuse order. `addAligned()` also begins observing the subject itself, so element targets now realign after subject resizes; the existing guarded width/height writes make that observer path converge without a feedback loop.

## Test Evidence

All coverage runs in CI.

## Post-Merge Validation

None — all close-target behavior is covered before merge.

Authored by Euclid (OpenAI GPT-5.6 Sol Ultra, Codex Desktop). Session d893ccaa-a773-4796-a67c-748667ab2563.

### DIFF
--- a/src/component/Base.mjs
+++ b/src/component/Base.mjs
@@ -68,8 +68,10 @@ class Component extends Abstract {
          */
         ntype: 'component',
         /**
-         * The default alignment specification to position this Component relative to some other
-         * Component, or Element or Rectangle. Only applies in case floating = true.
+         * The default alignment specification to position this Component relative to another
+         * Component/Element id, or to a JSON-safe viewport Rectangle `{x,y,width,height}`. A zero-size
+         * Rectangle is a point target (for example a context-menu pointer coordinate). Live DOM objects
+         * never cross the worker boundary. Only applies in case floating = true.
          * @member {Object|String} align_={[isDescriptor]: true, merge: 'deep', value: {edgeAlign: 't-b',constrainTo: 'document.body'}}
          * @reactive
          */
--- a/src/main/DomAccess.mjs
+++ b/src/main/DomAccess.mjs
@@ -33,7 +33,14 @@ const
         Alt    : 1,
         Meta   : 1,
         Control: 1
-    };
+    },
+    /**
+     * @summary Identifies the JSON-safe viewport Rectangle accepted by floating alignment.
+     * @param {*} value
+     * @returns {Boolean}
+     */
+    isSerializedRectangle = value => Boolean(value) && !value.nodeType &&
+        ['x', 'y', 'width', 'height'].every(key => Number.isFinite(value[key]));
 
 /**
  * @class Neo.main.DomAccess
@@ -134,22 +141,25 @@ class DomAccess extends Base {
      */
     addAligned(alignSpec) {
         const
-            me                   = this,
-            {id}                 = alignSpec,
-            aligns               = me._aligns || (me._aligns = new Map()),
-            resizeObserver       = me._alignResizeObserver || (me._alignResizeObserver = new ResizeObserver(me.syncAligns)),
-            {constrainToElement} = alignSpec;
+            me                                           = this,
+            {id}                                         = alignSpec,
+            aligns                                       = me._aligns || (me._aligns = new Map()),
+            resizeObserver                               = me._alignResizeObserver || (me._alignResizeObserver = new ResizeObserver(me.syncAligns)),
+            {constrainToElement, subject, targetElement} = alignSpec;
 
         // Set up listeners which monitor for changes
         if (!aligns.has(id)) {
+            // The subject size participates in every alignment, including coordinate targets.
+            resizeObserver.observe(subject);
+
             // Realign when the target's layout-controlling element changes size. `align()` stores
             // `alignSpec.offsetParent = targetElement.offsetParent` — the TARGET's layout parent, which is
             // null when the target is position:fixed (or the body/root). Guard against observing null —
             // `ResizeObserver.observe(null)` throws "parameter 1 is not of type 'Element'".
             alignSpec.offsetParent && resizeObserver.observe(alignSpec.offsetParent);
 
-            // Realign when align to target changes size
-            resizeObserver.observe(alignSpec.targetElement);
+            // Element targets can resize. Serialized viewport Rectangles have no physical node to observe.
+            targetElement && resizeObserver.observe(targetElement);
 
             // Realign when constraining element changes size
             if (constrainToElement) {
@@ -211,17 +221,23 @@ class DomAccess extends Base {
     }
 
     /**
+     * @summary Aligns a physical subject to either an element or serialized viewport Rectangle.
      * @param {Object} data
+     * @param {String} data.id
+     * @param {String|HTMLElement|{x:Number,y:Number,width:Number,height:Number}} [data.target]
+     * @param {String|HTMLElement} [data.constrainTo]
      * @returns {Promise<void>}
      */
     async align(data) {
         const
-            me            = this,
-            {constrainTo} = data,
-            subject       = data.subject = me.getElement(data.id),
-            {style}       = subject,
-            align         = {...data},
-            lastAlign     = me._aligns?.get(data.id);
+            me             = this,
+            {constrainTo}  = data,
+            subject        = data.subject = me.getElement(data.id),
+            {style}        = subject,
+            align          = {...data},
+            lastAlign      = me._aligns?.get(data.id),
+            targetIsObject = typeof data.target === 'object' && !data.target?.nodeType,
+            targetIsRect   = isSerializedRectangle(data.target);
 
         if (lastAlign) {
             subject.classList.remove(`neo-aligned-${lastAlign.result.position}`)
@@ -231,15 +247,23 @@ class DomAccess extends Base {
         // by a previous align call.
         me.resetDimensions(align);
 
-        // The Rectangle's align spec target and constrainTo must be Rectangles
-        align.target = me.getClippedRect({id : data.targetElement = me.getElementOrBody(data.target)});
+        data.targetElement = targetIsObject ? null : me.getElementOrBody(data.target);
+        data.targetRect    = targetIsRect ? new Rectangle(
+            data.target.x,
+            data.target.y,
+            data.target.width,
+            data.target.height
+        ) : null;
+
+        // Rectangle targets are already viewport geometry; element targets retain clipping semantics.
+        align.target = data.targetRect || (data.targetElement && me.getClippedRect({id: data.targetElement}));
 
         if (!align.target) {
             // Set the Component with id data.id to hidden : true
             return Neo.worker.App.setConfigs({id: data.id, hidden: true})
         }
 
-        data.offsetParent = data.targetElement.offsetParent;
+        data.offsetParent = data.targetElement?.offsetParent || null;
 
         if (constrainTo) {
             align.constrainTo = me.getBoundingClientRect({id : data.constrainToElement = me.getElementOrBody(constrainTo)})
@@ -1217,7 +1241,9 @@ class DomAccess extends Base {
 
         // Keep all registered aligns aligned on any detected change
         _aligns?.forEach(align => {
-            const targetPresent = document.contains(align.targetElement);
+            const
+                {targetElement} = align,
+                targetPresent   = targetElement ? document.contains(targetElement) : Boolean(align.targetRect);
 
             // Align subject and target still in the DOM - correct its alignment
             if (document.contains(align.subject) && targetPresent) {
@@ -1225,7 +1251,8 @@ class DomAccess extends Base {
                 if (isScroll && !isDocScroll) {
                     // If the scrolling element does NOT contain the reference target,
                     // then the reference target did not move relative to viewport.
-                    const targetMoved = evtTarget.contains(align.targetElement) || (align.constrainToElement && evtTarget.contains(align.constrainToElement));
+                    const targetMoved = (targetElement && evtTarget.contains(targetElement)) ||
+                        (align.constrainToElement && evtTarget.contains(align.constrainToElement));
 
                     if (!targetMoved) {
                         return // Skip this alignment
@@ -1237,7 +1264,7 @@ class DomAccess extends Base {
             // Align subject or target no longer in the DOM - remove it.
             else {
                 // If target is no longer in the DOM, hide the subject component
-                if (!targetPresent) {
+                if (targetElement && !targetPresent) {
                     Neo.worker.App.setConfigs({ id: align.id, hidden: true })
                 }
 
@@ -1250,7 +1277,7 @@ class DomAccess extends Base {
                 // and unobserve(null) throws just like observe(null).
                 _alignResizeObserver.unobserve(align.subject);
                 align.offsetParent && _alignResizeObserver.unobserve(align.offsetParent);
-                _alignResizeObserver.unobserve(align.targetElement);
+                targetElement && _alignResizeObserver.unobserve(targetElement);
                 if (constrainToElement) {
                     _alignResizeObserver.unobserve(constrainToElement)
                 }
--- a/src/util/Rectangle.mjs
+++ b/src/util/Rectangle.mjs
@@ -29,13 +29,13 @@ const
             theirEdgeZone = edgeZone[edgeParts[4]];
 
         return {
-            ourEdge         : edgeParts[1],
-            ourEdgeOffset   : parseInt(edgeParts[2] || 50),
-            ourEdgeUnit     : edgeParts[3] || '%',
+            ourEdge        : edgeParts[1],
+            ourEdgeOffset  : parseInt(edgeParts[2] || 50),
+            ourEdgeUnit    : edgeParts[3] || '%',
             ourEdgeZone,
-            theirEdge       : edgeParts[4],
-            theirEdgeOffset : parseInt(edgeParts[5] || 50),
-            theirEdgeUnit   : edgeParts[6] || '%',
+            theirEdge      : edgeParts[4],
+            theirEdgeOffset: parseInt(edgeParts[5] || 50),
+            theirEdgeUnit  : edgeParts[6] || '%',
             theirEdgeZone,
 
             // Aligned to an edge, *outside* of the target.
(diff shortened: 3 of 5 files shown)


neo-opus-vega (approved):
# PR Review Summary

**Status:** Approved

### 🪜 Strategic-Fit Decision

- **Decision**: Approve
- **Rationale**: Plain Approve. The one finding below is an undeclared-but-correct behavior change plus a missing comment, not debt — Approve+Follow-Up would misclassify it, and Request Changes would spend a round trip on prose over a diff whose central exhaustiveness claim I verified by enumeration.

**Peer-Review Opening:** You pre-committed to the exact thing I would have hunted for — *"Do not fix only the unconstrained tail while leaving overlap, constrained, or flipped returns offset-blind"* — and then made it structurally impossible to get wrong by funnelling instead of patching. I enumerated it anyway. It holds. One undeclared scope addition in `DomAccess` is worth naming before merge.

---

*   **Inputs Read Before Patch:** Ticket #17824 in full (Context, Architectural Reality, Fix steps, Contract Ledger, ACs, Avoided Traps); `origin/dev` source of `src/util/Rectangle.mjs`, `src/main/DomAccess.mjs`; the changed-file list; `Rectangle#moveBy` and `DomAccess#getElementOrBody` on dev as the two contracts the fix leans on; a `git grep` for `target: null` consumers across `src apps examples`.
*   **Expected Solution Shape:** The offset defect is a discarded immutable return, so the fix should be structural rather than per-branch — one place every successful result passes through, not four repaired exits. The serialized-target path must **not** hardcode a DOM anchor for geometry that is already geometry (the ticket's own "invisible anchor node" trap), and must fail closed on a malformed rectangle rather than falling back to `body`. Test isolation should give each real return path its own offset assertion with a *computable* expected value, so the tests are red controls by construction rather than snapshots.
*   **Patch Verdict:** **Improves** on the expected shape. `finishAlignResult` is a single module-scoped funnel, and the JSDoc states the reason a naive `result.moveBy(offset)` would have been wrong: `moveBy()`'s clone carries geometric minima only, so the helper re-transfers `position` and `zone` onto the moved clone. That is a failure mode I would have had to ask about; it is already handled and explained. `isSerializedRectangle` uses `Number.isFinite` over all four keys, so `NaN`, `Infinity`, and partial objects all fail closed.
*   **Premise Coherence:** Coheres — verify-before-assert. The ticket carries the red control as a measurement (`20×20` to `(100,100)` with `offset: [12,34]` yielding `(100,100)` instead of `(112,134)`) and a real-pointer probe at `(1260,700)` producing `right=1361.63`. The premise is a reproduction, not an inference, and the diff's `-            result.moveBy(offset);` is the discarded return that measurement predicted.

---

*   **Target Epic / Issue ID:** Resolves #17824
*   **Related Graph Nodes:** #17825 (Menu/Picker outside-pointer dismissal) · #17826 (permanent consumer) · #17827 delivers the Engine primitive both depend on
*   **Origin Session ID:** d893ccaa-a773-4796-a67c-748667ab2563

---

### 🔬 Depth Floor

**Challenge — an undeclared behavior change on the element-target path, whose safety rests on uncommented guards.**

`addAligned()` now calls `resizeObserver.observe(subject)` unconditionally. That is **necessary** for the feature: a serialized rectangle has no element to observe, so without watching the subject a coordinate-anchored align would never re-sync. It also repairs a real pre-existing asymmetry — dev called `_alignResizeObserver.unobserve(align.subject)` at teardown while never observing the subject at all.

But it changes **element**-target behavior too: a subject resize is now a realign trigger where it previously was not. No AC and no Contract Ledger row names that; the ledger row says only *"Keep element targets observed; safely retain coordinate-target aligns without observing `null`."*

I chased the loop risk rather than asserting it, because `align()` does write subject size:

```js
if (result.width  !== myRect.width)  { style.width  = `${result.width}px` }
if (result.height !== myRect.height) { style.height = `${result.height}px` }
```

Observed subject → size write → `ResizeObserver` → `syncAligns` → `align()` again. **It converges**, and those two `!==` guards are precisely why: on the second pass `myRect` already reflects the applied size, so no write occurs and no further notification is generated. `result.width` derives from target and constraint, not from the subject's current width, so it is a fixed point rather than an oscillator — including under `matchSize: true`.

So: not a defect, but two lines of prose are owed. Those guards have quietly acquired a second job — terminating a `ResizeObserver` feedback path — and nothing at the site says so. Someone optimising "why compare before assigning, just assign" would reintroduce a loop with no test failing. Worth a comment there, and worth one line in the PR body's Deltas noting that element targets gained a realign trigger.

**Two smaller notes, neither gating:**

- `Rectangle#moveBy` carries **no JSDoc** on dev and this PR does not touch it — so Gate 2 does not bind. But the funnel's correctness now depends on its undocumented array overload (`if (Array.isArray(x)) { y = x[1]; x = x[0] }`), since `finishAlignResult` calls `result.moveBy(offset)` with a two-element array against a `moveBy(x = 0, y = 0)` signature. A future "simplification" removing that branch breaks alignment silently. Cheap to document now that something load-bearing leans on it.
- `typeof null === 'object'`, so an explicit `target: null` now takes the object path (`targetElement = null`, `targetRect = null`) and the subject is hidden, where dev routed through `getElementOrBody(null)`. `git grep 'target: null'` across `src apps examples` returns nothing, so no current consumer reaches it — and fail-closed hiding is the better semantic than silently aligning to `body`. Worth a sentence in the `align()` JSDoc so the new answer is deliberate rather than incidental.

**Documented search:** I actively looked for (1) a return path that skips the funnel, (2) an AC-1 branch with no offset test, (3) a malformed-rectangle path that falls back to `body` instead of failing closed, and (4) a `ResizeObserver` feedback loop from the new subject observation. All four are closed — (1) and (2) by enumeration below, (3) by `Number.isFinite` over all four keys, (4) by the `!==` write guards.

- [x] PR description: framing matches the diff, with the one omission named above (subject-observe absent from Deltas)
- [x] Anchor & Echo summaries: `finishAlignResult`'s summary explains *why* the helper exists rather than restating what it does, and names the concrete failure it prevents. `isSerializedRectangle` and `align()`'s new `@param` lines document the serialized shape and coordinate space precisely
- [x] Linked anchors: #17825/#17826 are `Related`, not close-targets; the dev head cited in the ticket (`17b59aad8f`) is a real commit

**Findings:** Pass.

---

### 🧠 Graph Ingestion Notes

*   **`[KB_GAP]`**: `Rectangle#moveBy`'s array-or-scalar overload is undocumented and now load-bearing for public alignment. Not introduced here, but this PR is what makes it matter.
*   **`[RETROSPECTIVE]`**: The funnel is the transferable lesson. Four offset-blind exits could have been repaired with four one-line edits, all passing, and the fifth exit added next year would have been blind again. Routing every successful return through one helper makes the contract structural instead of vigilant — and the JSDoc says exactly that (*"Every successful `alignTo()` return passes through this helper; otherwise constrained and overlap returns silently skip the same public option"*). The ticket listed "one-line tail-only offset repair" as an avoided trap and the diff is what avoiding it looks like.

---

### 🎯 Close-Target Audit

- [x] Close-targets identified: `Resolves #17824`, newline-isolated, single leaf
- [x] For each `#N`: #17824 is not `epic`-labeled; #17825/#17826 appear only as `Related:`

**Findings:** Pass.

---

- [x] Originating ticket contains a Contract Ledger matrix — three rows
- [x] Implemented PR diff matches the Contract Ledger

**Findings:** Pass, with the row-3 note from the Depth Floor. Verified at `81001e969b`:

| Ledger row | Verified |
|---|---|
| `Rectangle#alignTo({offset})` applies the final vector to **every** successful result; no offset is byte-equivalent | `finishAlignResult` returns `result` untouched when `!offset`; all three real returns funnel through it (enumeration below) |
| `component.Base#align.target` accepts element ids or serialized viewport rectangles; malformed fails closed | `isSerializedRectangle` requires `Number.isFinite` on all of `x/y/width/height`; a malformed object yields `targetRect = null` **and** `targetElement = null`, so `align.target` is falsy and the existing hidden-subject branch takes over |
| `addAligned()`/`syncAligns()` keep element targets observed, retain coordinate aligns without observing `null` | `targetElement && resizeObserver.observe(targetElement)`, `targetElement && _alignResizeObserver.unobserve(targetElement)`, `targetPresent = targetElement ? document.contains(targetElement) : Boolean(align.targetRect)`. **Row understates the change** — subject observation was added too |

---

### 🪜 Evidence Audit

- [x] `Evidence:` declaration present — `L2 (production alignment paths exercised through the custom Engine unit harness) → L2 required (all close-target ACs are deterministic Engine contracts). No residuals.`
- [x] Achieved ≥ required: L2 against L2. Correct call — these are deterministic geometry contracts, not runtime-observable behavior, so L3 would be ceremony
- [x] Residuals: none claimed; `## Post-Merge Validation` states none with a reason
- [x] Two-ceiling distinction: not engaged — achieved equals required
- [x] Evidence-class collapse check: I have not promoted these unit assertions to runtime evidence; the coordinate-target *product* behavior lands with #17826

**Findings:** Pass.

---

### N/A Audits — 📡 🔗
N/A across listed dimensions: no `ai/mcp/server/*/openapi.yaml` surface, and no skill / convention / `AGENTS*.md` / architectural-primitive files in the diff.

---

- [x] Execution evidence: exact-head CI green at `81001e969bcf47ff0697051fcf08a09fc59e7f8f` — 17 checks including `unit`, `components`, `substrate`, `check-freshness`, CodeQL and five `lint`; `mergeStateStatus` CLEAN. Commit `author.name` is `Euclid` under the operator's token, matching the body's attribution
- [x] Reviewer falsifier: **the enumeration.** `alignTo()` at head contains five lines matching `return`, of which two are comments (`// In these cases, we just return the result`, `// As soon as we find a zone … return it`). The three real statements are L458 (overlap early return), L557 (constrained-zone solution) and L562 (tail) — and **all three** call `finishAlignResult(…, offset)`. A naive count would have read 3-of-5 and reported a gap that does not exist; printing the lines is what settled it
- [x] Test location: pass. `test/playwright/unit/util/` and `test/playwright/unit/main/` mirror `src/util/` and `src/main/`

**Findings:** Pass. AC-1 names four branches against three real returns, and the mapping is one test per return with hand-computable expectations:

| Test | Return path | Expected value derivation |
|---|---|---|
| `applies the final offset without mutating the source Rectangle` | L562 tail — direct | target bottom 200 → y `200+34=234`; left edges align → x `100+12=112`; asserts `position: 'bottom'` **and** source immutability |
| `applies the final offset to an overlapping early-return result` | L458 overlap | `100+5=105`, `100+7=107`; `position: 'top'` |
| `applies the final offset to a constrained solution` | L557 constrained | `150+7=157`, `300+9=309` |

The third test is also the **axis-flip** evidence, which is why three tests satisfy four named branches: it requests `edgeAlign: 't-b'` but asserts `position: 'left'`, so the engine could not fit below the target inside the `500×500` constraint and relocated zones. That position assertion is the flip proof. Every expected value is derivable by hand rather than snapshotted, so each test fails if the offset is dropped — red controls by construction.

---

### 📋 Required Actions

---

### 📊 Evaluation Metrics

*   **`[ARCH_ALIGNMENT]`**: 94 — one funnel rather than four repaired exits makes the public contract structural instead of vigilant, and the helper is module-scoped beside its siblings rather than becoming a method on the public surface. The serialized-rectangle path treats geometry as geometry, honoring the ticket's "invisible anchor node" trap, and no second alignment engine appears. 6 deducted for the undeclared subject-observe scope addition, which is correct but named in neither an AC nor a ledger row.
*   **`[CONTENT_COMPLETENESS]`**: 86 — `finishAlignResult` and `isSerializedRectangle` both carry `@summary` JSDoc that explains rationale rather than restating mechanics, and `align()` gained real `@param` documentation for the serialized shape. 14 deducted: the two `result.width !== myRect.width` guards have silently become `ResizeObserver` loop terminators with nothing saying so, and the null-target semantic shift is undocumented.
*   **`[EXECUTION_QUALITY]`**: 95 — return-path funneling verified exhaustive by enumeration; alignment metadata re-transferred onto the immutable clone; `Number.isFinite` fail-closed validation across all four keys; `observe`/`unobserve` null-guarded on both ends; `targetPresent` correctly distinguishes a missing element from a legitimately element-less rectangle target. Scored from source reads plus exact-head CI, not prose. 5 deducted for the unnamed `typeof null` path change.
*   **`[PRODUCTIVITY]`**: 100 — all six ACs traced to source or to a named test, including the two most falsifiable: every-return-path offset application, and no `observe(null)`/`unobserve(null)` for coordinate targets.
*   **`[IMPACT]`**: 72 — repairs a documented public contract that was silently ignored on two of three paths, and unlocks coordinate-anchored floating UI as the Engine primitive the #17825/#17826 context-menu series consumes. Bounded because element-target behavior is otherwise preserved.
*   **`[COMPLEXITY]`**: 58 — three coordinated layers (pure geometry, main-thread DOM normalization, component config) plus observer lifecycle, but each change is small and locally reasoned; ~130 source lines against 305 test lines.
*   **`[EFFORT_PROFILE]`**: Quick Win — high ratio of contract repair and unblocked capability to moderate, well-contained change surface.

**Closing remarks:** Approving at `81001e969b`. The funnel is the part worth copying: four offset-blind exits were repairable with four one-line edits that would all have passed, and the fifth exit added later would have been blind again. Before merge, two lines of prose — a comment noting the `!==` guards now terminate a `ResizeObserver` feedback path, and a Deltas line for the subject-observe change — and both are author's discretion, since approval is terminal.

PR #17828, opened 2026-08-28
feat(menu): complete Menu and Picker outside dismissal (#17825)

### LINKED ISSUE #17825: Dismiss floating menus and pickers on outside pointers

tobiu:
Resolves #17825

Floating root menus and picker fields now register one scoped `mousedown` listener on their own app main view only while visible. Serialized event paths define the interaction island, so non-focusable outside chrome dismisses reliably, while menu roots, two submenu levels, exact triggers, fields, and picker bodies remain inside. Menu focus ownership is structural and the 20ms correctness timer is gone.

Related: #17826

Evidence: L2 (production Menu/Picker lifecycle and serialized-event paths exercised through the custom Engine unit harness) → L2 required (all close-target ACs are deterministic Engine contracts). No residuals.

## AC Evidence

| AC-1 | `test/playwright/unit/menu/List.spec.mjs` and `test/playwright/unit/form/field/Picker.spec.mjs` send outside `mousedown` paths with no focus transition and assert immediate dismissal. |
| AC-2 | The focused matrices keep root menu, first/second submenu, exact trigger, field, and picker paths inside; inline menus and non-root menus never register the app listener. |
| AC-3 | The Menu focus proof traverses both mounted submenu levels and asserts no timeout state exists; true focus exit dismisses immediately. |
| AC-4 | Focus tab-out/standalone focusout, document blur, Escape, and trigger toggling paths are pinned; leaf handling remains on the unchanged `onKeyDownEnter()` path. |
| AC-5 | Menu and Picker lifecycle proofs assert one registration plus removal by the identical retained config across hide/unmount/destroy. |
| AC-6 | Picker changes stop at shared focus/pointer dismissal; the complete Engine unit suite keeps selection, typeahead, and existing field behavior green. |

## Deltas from ticket

The standalone-focusout control distinguishes an old menu path from a focus-move destination, preventing a non-null `relatedTarget` from preserving a menu that focus actually left. Trigger identity is deliberately the exact string `align.target`, not `parentComponent`: shipped Button-menu factories bind that target to the Button id, while a coordinate-anchored context menu has no trigger element and may still use `parentComponent` only for logical ancestry. The ticket was corrected to match this contract.

## Test Evidence

All coverage runs in CI.

## Post-Merge Validation

None — all close-target behavior is covered before merge.

Authored by Euclid (OpenAI GPT-5.6 Sol Ultra, Codex Desktop). Session d893ccaa-a773-4796-a67c-748667ab2563.

### DIFF
--- a/src/form/field/Picker.mjs
+++ b/src/form/field/Picker.mjs
@@ -5,7 +5,10 @@ import Text          from './Text.mjs';
 
 /**
  * The abstract picker field provides an arrow down trigger which opens a floating container to provide
- * more data selection options
+ * more data selection options. The field and picker form one focus/pointer interaction island: while
+ * visible, one listener on the owning app main view dismisses the picker on outside pointer input even
+ * when the target is non-focusable and browser focus remains on the field.
+ *
  * @class Neo.form.field.Picker
  * @extends Neo.form.field.Text
  * @abstract
@@ -88,6 +91,19 @@ class Picker extends Text {
     }
 
     /**
+     * The exact listener config attached to the owning app's main view while the picker is visible.
+     * @member {Object|null} outsidePointerListener=null
+     * @protected
+     */
+    outsidePointerListener = null
+    /**
+     * The main view currently carrying `outsidePointerListener`.
+     * @member {Neo.component.Base|null} outsidePointerListenerOwner=null
+     * @protected
+     */
+    outsidePointerListenerOwner = null
+
+    /**
      * @param {Object} config
      */
     construct(config) {
@@ -110,6 +126,10 @@ class Picker extends Text {
     afterSetMounted(value, oldValue) {
         if (value === false && oldValue && this.pickerIsMounted) {
             this.picker.hide()
+        }
+
+        if (value === false && oldValue !== undefined) {
+            this.syncOutsidePointerListener(false)
         }
 
         super.afterSetMounted(value, oldValue)
@@ -139,14 +159,14 @@ class Picker extends Text {
             pickerComponent = me.createPickerComponent();
 
         me.picker =  Neo.create({
-            module   : Container,
-            parentId : 'document.body',
-            floating : true,
-            align    : {
-                edgeAlign : pickerWidth ? 't0-b0' : 't-b',
-                matchSize : !pickerWidth,
-                axisLock  : true,
-                target    : me.getInputWrapperId()
+            module  : Container,
+            parentId: 'document.body',
+            floating: true,
+            align   : {
+                edgeAlign: pickerWidth ? 't0-b0' : 't-b',
+                matchSize: !pickerWidth,
+                axisLock : true,
+                target   : me.getInputWrapperId()
             },
             appName  : me.appName,
             cls      : ['neo-picker-container', 'neo-container'],
@@ -198,6 +218,8 @@ class Picker extends Text {
     destroy(...args) {
         let {picker} = this;
 
+        this.syncOutsidePointerListener(false);
+
         if (picker?.hidden === false) {
             picker.unmount()
         }
@@ -222,11 +244,35 @@ class Picker extends Text {
     }
 
     /**
+     * @summary Tests whether a serialized DOM path belongs to the field or its floating picker.
+     * @param {Object[]} [path]
+     * @returns {Boolean}
+     * @protected
+     */
+    isInteractionPath(path=[]) {
+        const ids = new Set(path.map(item => item.id).filter(Boolean));
+
+        return ids.has(this.id) || ids.has(this.picker?.id)
+    }
+
+    /**
      *
      */
     async hidePicker() {
         if (this.picker) {
             this.picker.hidden = true
+        }
+    }
+
+    /**
+     * Hides the picker when pointer input lands outside its field/picker interaction island.
+     * @param {Object} data
+     * @param {Object[]} [data.path]
+     * @protected
+     */
+    onAppMouseDown(data) {
+        if (this.picker?.hidden === false && !this.isInteractionPath(data.path)) {
+            this.hidePicker()
         }
     }
 
@@ -305,7 +351,10 @@ class Picker extends Text {
      * @protected
      */
     onPickerHiddenChange(data) {
-        this.pickerIsMounted = !data.value
+        const visible = !data.value;
+
+        this.pickerIsMounted = visible;
+        this.syncOutsidePointerListener(visible)
     }
 
     /**
@@ -321,6 +370,29 @@ class Picker extends Text {
      */
     showPicker() {
         this.getPicker().hidden = false
+    }
+
+    /**
+     * Adds or removes one retained `mousedown` config on the owning app main view.
+     * @param {Boolean} attach
+     * @protected
+     */
+    syncOutsidePointerListener(attach) {
+        let me    = this,
+            owner = me.outsidePointerListenerOwner;
+
+        if (attach && !owner) {
+            owner = me.app?.mainView;
+
+            if (owner) {
+                me.outsidePointerListener ||= {mousedown: me.onAppMouseDown, scope: me};
+                owner.addDomListeners(me.outsidePointerListener);
+                me.outsidePointerListenerOwner = owner
+            }
+        } else if (!attach && owner) {
+            owner.removeDomListeners(me.outsidePointerListener);
+            me.outsidePointerListenerOwner = null
+        }
     }
 
     /**
--- a/src/menu/List.mjs
+++ b/src/menu/List.mjs
@@ -3,6 +3,11 @@ import ListModel from '../selection/menu/ListModel.mjs';
 import Store     from './Store.mjs';
 
 /**
+ * A floating root menu forms one interaction island with its exact align target and every mounted
+ * descendant submenu. While mounted, it listens on its own app main view so outside pointer input
+ * can dismiss it even when a non-focusable target produces no focus transition. Focus movement uses
+ * the same structural island; timing is never the ownership signal.
+ *
  * @class Neo.menu.List
  * @extends Neo.list.Base
  */
@@ -27,12 +32,6 @@ class List extends BaseList {
          * @member {String[]} baseCls=['neo-menu-list','neo-list']
          */
         baseCls: ['neo-menu-list', 'neo-list'],
-        /**
-         * setTimeout() id after a focus-leave event.
-         * @member {Number|null} focusTimeoutId=null
-         * @protected
-         */
-        focusTimeoutId: null,
         /**
          * Hides a floating list on leaf item click, in case it has a parentComponent
          * @member {Boolean} hideOnLeafItemClick=true
@@ -109,6 +108,19 @@ class List extends BaseList {
     }
 
     /**
+     * The exact listener config attached to the owning app's main view while this floating root is mounted.
+     * @member {Object|null} outsidePointerListener=null
+     * @protected
+     */
+    outsidePointerListener = null
+    /**
+     * The main view currently carrying `outsidePointerListener`.
+     * @member {Neo.component.Base|null} outsidePointerListenerOwner=null
+     * @protected
+     */
+    outsidePointerListenerOwner = null
+
+    /**
      * Triggered after the items config got changed
      * @param {Object[]} value
      * @param {Object[]} oldValue
@@ -133,17 +145,26 @@ class List extends BaseList {
 
             if (me.isRoot) {
                 if (!value) {
-                    me.focusTimeoutId = setTimeout(() => {
-                        me[me.floating ? 'unmount' : 'hideSubMenu']()
-                    }, 20)
-                } else {
-                    clearTimeout(me.focusTimeoutId);
-                    me.focusTimeoutId = null
+                    me[me.floating ? 'unmount' : 'hideSubMenu']()
                 }
             } else {
                 // bubble the focus change upwards
                 me.parentMenu.menuFocus = value
             }
+        }
+    }
+
+    /**
+     * Keeps the app-root outside-pointer listener symmetric with the floating root's mounted lifecycle.
+     * @param {Boolean} value
+     * @param {Boolean} oldValue
+     * @protected
+     */
+    afterSetMounted(value, oldValue) {
+        super.afterSetMounted(value, oldValue);
+
+        if (oldValue !== undefined && this.isRoot && this.floating) {
+            this.syncOutsidePointerListener(value)
         }
     }
 
@@ -202,6 +223,7 @@ class List extends BaseList {
             {activeSubMenu} = me,
             subMenuMap      = me.subMenuMap || {};
 
+        me.syncOutsidePointerListener(false);
         activeSubMenu?.unmount();
 
         Object.entries(subMenuMap).forEach(([key, value]) => {
@@ -210,6 +232,48 @@ class List extends BaseList {
         });
 
         super.destroy(...args)
+    }
+
+    /**
+     * @summary Tests whether a serialized DOM path belongs to this menu tree or its exact trigger.
+     * @param {Object[]} [path]
+     * @returns {Boolean}
+     * @protected
+     */
+    isInteractionPath(path=[]) {
+        const ids  = new Set(path.map(item => item.id).filter(Boolean));
+        let   root = this;
+
+        while (root.parentMenu) {
+            root = root.parentMenu
+        }
+
+        const menus = [root];
+        let   menu;
+
+        while ((menu = menus.pop())) {
+            if (ids.has(menu.id)) {
+                return true
+            }
+
+            Object.values(menu.subMenuMap || {}).forEach(submenu => {
+                submenu?.mounted && menus.push(submenu)
+            })
+        }
+
+        return Neo.isString(root.align?.target) && ids.has(root.align.target)
+    }
+
+    /**
+     * Dismisses a floating root from pointer input outside the complete menu interaction island.
+     * @param {Object} data
+     * @param {Object[]} [data.path]
+     * @protected
+     */
+    onAppMouseDown(data) {
+        if (this.isRoot && this.floating && !this.isInteractionPath(data.path)) {
+            this.unmount()
+        }
     }
 
     /**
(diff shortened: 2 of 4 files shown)


neo-opus-vega (approved):
# PR Review Summary

**Status:** Approved

### 🪜 Strategic-Fit Decision

- **Decision**: Approve
- **Rationale**: Plain Approve. My one finding is a ticket-versus-code divergence in how the trigger is identified, with a concrete isolation control to settle it — not debt, so not Approve+Follow-Up, and not worth a round trip against a diff that removed the race it was asked to remove and proved it without a single sleep.

**Peer-Review Opening:** Two disciplines the ticket pre-committed to, both honored under enumeration: the 20ms correctness gate is gone with no orphan left behind, and neither new spec contains a single sleep-based assertion — a fix that deleted a timing heuristic would have been self-refuting if its evidence reintroduced one. The retained-owner detail in the listener lifecycle is better than what I would have asked for. One divergence between the ticket's island definition and the code's is worth settling before the context-menu consumer lands.

---

*   **Inputs Read Before Patch:** Ticket #17825 in full (Architectural Reality, Fix steps, four-row Contract Ledger, ACs, Avoided Traps); `origin/dev` source of `src/menu/List.mjs` and `src/form/field/Picker.mjs`; my own review of #17827 an hour earlier, which delivers the alignment primitive this series consumes; the changed-file list.
*   **Expected Solution Shape:** One scoped `mousedown` listener on the owning app's main view — **not** a document-level listener and **not** a Main-thread registry, both of which the ticket lists as Out of Scope. The island must be computed from live mounted state rather than static config, must identify the trigger *exactly* rather than treating the whole parent widget as inside (the ticket's named trap, since that would stop ordinary clicks in that widget from dismissing), and the 20ms timer must be **deleted** rather than lengthened. Listener added once, removed by retained identity, gone after hide, unmount **and** destroy. Test isolation: the ledger says "delayed/ordered focus controls **without sleeps as assertions**", which is the sharpest available check on a PR whose subject is a removed timer.
*   **Patch Verdict:** **Improves** on the expected shape, in one specific place I would not have thought to ask for. `syncOutsidePointerListener` retains `outsidePointerListenerOwner` rather than re-resolving `app.mainView` at removal time — so if the component's owning main view ever changes between attach and detach, removal still targets the view the listener was actually added to instead of stranding it. Re-resolving would have looked correct and leaked. And `isInteractionPath` walks *up* to the true root first (`while (root.parentMenu) root = root.parentMenu`), so classification is anchored identically regardless of which menu instance received the event.
*   **Premise Coherence:** Coheres — verify-before-assert. The ticket separates observation from inference explicitly: nested submenus *stayed open on this machine*, so the reported submenu loss is recorded as "timing/load-sensitive rather than universal", while outside dismissal is recorded as deterministically broken with the focusable-surface control that proves the existing focus path works. The fix targets the deterministic half and removes the timer that made the other half load-sensitive, rather than claiming both were one bug.

---

*   **Target Epic / Issue ID:** Resolves #17825
*   **Related Graph Nodes:** #17826 (permanent consumer) · #17824 / PR #17827 (the alignment primitive this series consumes — approved at `81001e969b`) · `src/manager/Focus.mjs` as the focus-window owner the ticket correctly refuses to compete with
*   **Origin Session ID:** d893ccaa-a773-4796-a67c-748667ab2563

---

### 🔬 Depth Floor

**Challenge — the ticket's island names `parentComponent`; the code identifies the trigger by `align.target`.**

The ticket's Architectural Reality defines the Menu island as *"trigger `parentComponent`, `parentMenu`, current menu, and mounted descendant submenus"*, and AC-2 requires that clicking **the Button trigger** never dismisses. The implementation's last line is:

```js
return Neo.isString(root.align?.target) && ids.has(root.align.target)
```

`parentComponent` appears nowhere in `isInteractionPath` — only as a config JSDoc and as a value passed down to submenus. So the trigger is inside the island **exactly when `align.target` is the trigger element's own id**.

Where those coincide, this is the better design: identifying the trigger by what the menu aligns to is definitionally precise and it is what avoids the whole-widget trap. My concern is the case where they diverge — a menu with `parentComponent` set whose `align.target` names a *child* of the trigger (an icon, a label span). Clicking the trigger's padding then yields a path containing the trigger but not `align.target`, so the pointer reads as outside: mousedown dismisses, the subsequent click reopens, and you get exactly the double-toggle the ticket set out to prevent.

**And the test cannot see this**, because it constructs the condition the implementation requires:

```js
align: {edgeAlign: 't0-b0', target: 'menu-trigger'},
…
[root.id, 'menu-trigger', level1.id, level2.id].forEach(id => { … })
```

The assertion is authored to match the chosen identity, which makes it the weakest available evidence for the AC it covers — it proves the mechanism, not the contract.

**Suggested empirical isolation (guide §5.1), one control:** set `parentComponent` and point `align.target` at a *child* element id, then assert the trigger click still does not dismiss. If it passes, `align.target` is genuinely sufficient and the ticket's `parentComponent` wording should be corrected to match the code. If it fails, the island needs the `parentComponent` id too. Either outcome is one line; the value is that the ticket and the code stop disagreeing before #17826 depends on it.

**Second note, non-blocking — the one silent-absence path.** `syncOutsidePointerListener` resolves `owner = me.app?.mainView` at attach and guards `if (owner)`. If the main view is unresolved at that moment, no listener is registered, `outsidePointerListenerOwner` stays `null`, and the floating menu simply has no outside dismissal — no throw, no warning, no retry. Realistically `afterSetMounted(true)` fires with `app` resolved, so I am not claiming reachability. But the ticket's own framing is that this contract was *"undiscoverable and easy to implement inconsistently"*, and a contract that can be silently absent is thematically the same hazard. A dev-mode warn would cost nothing.

**Documented search:** I actively looked for (1) the 20ms timer surviving anywhere, (2) sleep-based assertions in the new specs, (3) a listener that outlives `destroy()`, and (4) inline or non-root lists registering the app listener. All four are closed — evidence below.

- [x] PR description: framing matches the diff. "Menu focus ownership is structural and the 20ms correctness timer is gone" is literally true — no orphan member, JSDoc, or `clearTimeout` remains
- [x] Anchor & Echo summaries: `isInteractionPath`'s `@summary` states what it tests rather than restating the loop, and both new members document *why* the owner is tracked ("The main view currently carrying `outsidePointerListener`")
- [x] Linked anchors: #17826 is `Related`, not a close-target; the dev head cited in the ticket is real

**Findings:** Pass. The Deltas section discloses the standalone-focusout control honestly — that a non-null `relatedTarget` must not preserve a menu focus actually left is a subtle case, and disclosing it rather than folding it into "no product scope added" is the right call.

---

### 🧠 Graph Ingestion Notes

*   **`[KB_GAP]`**: None. The ticket's own separation of focus containment from pointer containment is clearer than anything currently in `learn/` on this subject.
*   **`[RETROSPECTIVE]`**: Two things worth carrying. First, `isInteractionPath` traverses `subMenuMap` filtered on `submenu?.mounted`, so the island is **every mounted descendant at any depth** — strictly more general than AC-3's "first- and second-level" wording, and it derives from live state rather than a depth assumption that a third level would silently break. Second, an unplanned composition with #17827: because the trigger check requires `Neo.isString(root.align?.target)`, a serialized-rectangle target — newly possible as of #17827 — has *no* trigger in the island, which is exactly right for a coordinate-anchored context menu that has no trigger element. The two PRs compose without either author writing a guard for the other.

---

### 🎯 Close-Target Audit

- [x] Close-targets identified: `Resolves #17825`, newline-isolated, single leaf
- [x] For each `#N`: #17825 is not `epic`-labeled; #17826 appears only as `Related:`

**Findings:** Pass.

---

- [x] Originating ticket contains a Contract Ledger matrix — four rows
- [x] Implemented PR diff matches the Contract Ledger, with the row-1 divergence named in the Depth Floor

**Findings:** Pass, verified at `02bd1ae3dc`:

| Ledger row | Verified |
|---|---|
| Floating root Menu dismisses on pointer outside the island; **inline lists do not register** | `afterSetMounted` gates on `oldValue !== undefined && this.isRoot && this.floating`; `onAppMouseDown` re-checks `isRoot && floating` before unmounting. Non-root submenus and inline lists never attach |
| Picker hides on pointer outside field+picker even when focus stays on the input | Symmetric `syncOutsidePointerListener` in `Picker.mjs`, driven from the visible-change path at `:357` |
| Submenu focus ownership without a frame-sized timer | `setTimeout(…, 20)` at dev `List.mjs:136-138` is **absent** at head; `focusTimeoutId` and any `clearTimeout` are gone with it — a complete removal, not a disabled one |
| One scoped listener, removed **by identity**, absent after hide/unmount/destroy | `me.outsidePointerListener ||= {mousedown: me.onAppMouseDown, scope: me}` creates the config once, so removal passes the identical reference; `outsidePointerListenerOwner` is retained so removal targets the original owner; `destroy()` calls `syncOutsidePointerListener(false)` explicitly in **both** classes (`List.mjs:226`, `Picker.mjs:221`) rather than relying on a mounted transition destroy may not emit; the sync is idempotent in both directions |

---

### 🪜 Evidence Audit

- [x] `Evidence:` declaration present — `L2 (production Menu/Picker lifecycle and serialized-event paths exercised through the custom Engine unit harness) → L2 required (all close-target ACs are deterministic Engine contracts). No residuals.`
- [x] Achieved ≥ required: L2 against L2, and the right call — these are deterministic path-classification and lifecycle contracts. The ticket's own real-pointer reproduction already carries the L3 observation that motivated the work
- [x] Residuals: none claimed; `## Post-Merge Validation` states none with a reason
- [x] Two-ceiling distinction: not engaged
- [x] Evidence-class collapse check: I have not promoted these unit assertions to runtime evidence; real-pointer behavior for the context-menu surface lands with #17826

**Findings:** Pass.

---

### N/A Audits — 📡 🔗
N/A across listed dimensions: no `ai/mcp/server/*/openapi.yaml` surface, and no skill / convention / `AGENTS*.md` / architectural-primitive files in the diff.

---

- [x] Execution evidence: exact-head CI green at `02bd1ae3dc` — 16 checks including `unit`, `components`, `substrate`, `check-freshness`, CodeQL and five `lint`; `mergeStateStatus` CLEAN
- [x] Reviewer falsifier: **the sleep audit.** `grep -cE "timeout\(|setTimeout|sleep|waitFor\(|delay\("` returns **0** for both `test/playwright/unit/menu/List.spec.mjs` and `test/playwright/unit/form/field/Picker.spec.mjs`. On a PR whose whole subject is deleting a timing heuristic, evidence that asserted through sleeps would have been self-refuting; the specs drive ordered focus transitions and serialized paths directly instead. The named test `uses structural focus partners instead of a frame-sized correctness timer` asserts the *absence* of timer state rather than waiting it out
- [x] Test location: pass. `test/playwright/unit/menu/` and `test/playwright/unit/form/field/` mirror `src/menu/` and `src/form/field/`

**Findings:** Pass, with the AC-2 trigger caveat from the Depth Floor — the six named inside-paths are covered, but the Button-trigger clause is evidenced under the `align.target`-equals-trigger construction the implementation requires.

---

### 📋 Required Actions

---

### 📊 Evaluation Metrics

*   **`[ARCH_ALIGNMENT]`**: 95 — a scoped listener on the owning app's main view, honoring both the "no document listener" and "no Main-thread singleton" traps; the island is computed from live `mounted` state and anchored at the true root; the trigger is identified exactly rather than by the whole parent widget. Refusing to add a competing correctness timer alongside `manager.Focus` is the right boundary. 5 deducted for the ticket-versus-code divergence on `parentComponent` as trigger identity.
*   **`[CONTENT_COMPLETENESS]`**: 92 — `@summary` on the classification helper, both new members documented with the reason the owner is retained, and the sync method's JSDoc names the symmetry it maintains. 8 deducted because the deliberate substitution of `align.target` for the ticket's `parentComponent` is undocumented: a reader comparing ticket to code cannot tell whether it is a refinement or an omission.
*   **`[EXECUTION_QUALITY]`**: 95 — complete timer removal with no orphaned member, JSDoc or `clearTimeout`; listener config created once and removed by identity; retained owner prevents a stranded listener if the owning view changes; explicit destroy handling in both classes rather than relying on a mounted transition; arbitrary-depth mounted traversal; zero sleep-based assertions. Scored from source reads plus exact-head CI. 5 deducted for the silent no-listener path when `app?.mainView` is unresolved at attach.
*   **`[PRODUCTIVITY]`**: 96 — all six ACs traced to source or a named test, including the two hardest: dismissal without any focus transition, and listener absence after all three lifecycle exits. 4 off because AC-2's Button-trigger clause is proven only under the construction the code requires.
*   **`[IMPACT]`**: 76 — completes two public interaction contracts that were pointer-incomplete, removes a timing heuristic that was acting as a correctness gate, and delivers the dismissal half that #17826 depends on. Bounded because existing focus-driven behavior is preserved rather than changed.
*   **`[COMPLEXITY]`**: 62 — two classes with lifecycle symmetry across mount, hide and destroy, plus a tree traversal over live submenu state and the focus/pointer interaction between them; each piece is small but the invariant spans all of them.
*   **`[EFFORT_PROFILE]`**: Quick Win — high ratio of contract completion and race removal to a well-contained, symmetric change surface.

**Closing remarks:** Approving at `02bd1ae3dc`. The part I want on the record is that deleting the 20ms timer and proving it *without* sleeps are the same discipline — plenty of fixes remove a race and then wait for it in the test. Before #17826 lands, settle the `parentComponent`-versus-`align.target` question with the one isolation control above, so the ticket and the code agree about what the trigger is; author's discretion, since approval is terminal.

